PR #91, opened 2016-05-18
issue: 777358 un-offload UDP socket that is bound to any non-offloaded interfaces

rosenbaumalex:
(like loopback (127.0.0.1)

In case we detect sockets are bound to not offloaded interfaces we
should remove them from the offloaded list (remove from fd_colelction).
Calling setPassthrough() in bind() flow will handle that.

If this not-offloaded socket is called from select(), it will not be
mapped as offloaded and select(0 will refrain from checking the RING's
for events.
Also, the vma_stats will not see these fd's as offloaded.


mellanox-github:
Test PASSed.
See http://hpc-master.lab.mtl.com:8080/job/accl-libvma-pr/89/ for details (Mellanox internal link).

PR #95, opened 2016-05-20
issue: 777358 un-offload UDP socket that is bound to any non-offloaded interfaces (like loopback (127.0.0.1)

rosenbaumalex:
In case we detect sockets are bound to not offloaded interfaces we
should remove them from the offloaded list (remove from fd_colelction).
Calling setPassthrough() in bind() flow will handle that.

If this not-offloaded socket is called from select(), it will not be
mapped as offloaded and select(0 will refrain from checking the RING's
for events.
Also the vma_stats will not see these fd's as offloaded.

### DIFF
--- a/src/vma/sock/sockinfo_udp.cpp
+++ b/src/vma/sock/sockinfo_udp.cpp
@@ -474,7 +474,7 @@ int sockinfo_udp::bind(const struct sockaddr *__addr, socklen_t __addrlen)
 	}
 
 	struct sockaddr_in bound_addr;
-	socklen_t   boundlen = sizeof(struct sockaddr_in);
+	socklen_t boundlen = sizeof(struct sockaddr_in);
 	struct sockaddr *name = (struct sockaddr *)&bound_addr;
 	socklen_t *namelen = &boundlen;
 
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test PASSed.
See http://hpc-master.lab.mtl.com:8080/job/accl-libvma-pr/98/ for details (Mellanox internal link).

src/vma/sock/sockinfo_udp.cpp
```diff
@@ -672,18 +672,19 @@ int sockinfo_udp::on_sockname_change(struct sockaddr *__name, socklen_t __namele
 	if (is_bound_modified && bound_port != INPORT_ANY) {
 
 		// Attach UDP unicast port to offloaded interface
-		// 1. Verify not binding to MC address in the UC case
```
OphirMunk: @rosenbaumalex referencing line 664
```code
if (m_bound.get_in_addr() != bound_if) {
	si_udp_logdbg("bound if changed (%s -> %d.%d.%d.%d)", m_bound.to_str_in_addr(), NIPQUAD(bound_if));
	m_bound.set_in_addr(bound_if);
	m_p_socket_stats->bound_if = bound_if;
        is_bound_modified = true;                                  <== this line is *NOT* in the original code. Isn't it required?
}
```
Please note question in code

src/vma/sock/sockinfo_udp.cpp
```diff
@@ -672,18 +672,19 @@ int sockinfo_udp::on_sockname_change(struct sockaddr *__name, socklen_t __namele
...
+		}
 		else {
-			if (m_bound.is_mc()) {
-				si_udp_logdbg("bound to MC address, no need to attach to UC address as offloaded");
-			}
-			else {
-				si_udp_logdbg("will be passed to OS for handling - not offload interface (%s)", m_bound.to_str());
-			}
+			si_udp_logdbg("will be passed to OS for handling - not offload interface (%s)", m_bound.to_str());
+			setPassthrough();
```
OphirMunk: @rosenbaumalex In commit message you indicate that setPassthrough() is handling the removal a socket from fd_collection. When inspecting setPassthrough() body socket removal is not seen. Does it happen at a later phase?
Is the message valid?

rosenbaumalex: this is handled in https://github.com/Mellanox/libvma/blob/master/src/vma/sock/sock-redirect.cpp#L565


```
ret = p_socket_object->bind(__addr, __addrlen);
		if (p_socket_object->isPassthrough()) {
			handle_close(__fd, false, true);
			if (ret) {
				ret = orig_os_api.bind(__fd, __addr, __addrlen);
			}
		}

```


after the call to the sockinfo::bind() which will set the PassThrough flag as ON, the sock-redirect will logically close the socket from the fd_collection and pass it to OS for further handling.

src/vma/sock/sockinfo_udp.cpp
rosenbaumalex: We want to create HW flow steering rules only once we have a L4 port defined. the IP address will be the unicats interface address or the MC address.
So the code only verifies the bound PORT != 0, so we can proceed with HW steering definitions.

rosenbaumalex:
let's wait with merge for after VMA-8.0 is formally released

PR #96, opened 2016-05-20
issue: 781540 Add short (25 msec) delay between RAW QP state change to ERROR and polling CQ for all CQE's

rosenbaumalex:
In ConnectX-4 we need this additional delay allow all Rx WQE's to
complete FLUSH-ing into CQ before closing the QP and CQ.
This can cause VMA Rx buffer resource leak when closing a RING.

### DIFF
--- a/src/vma/dev/ring_simple.cpp
+++ b/src/vma/dev/ring_simple.cpp
@@ -133,6 +133,9 @@ ring_simple::~ring_simple()
 	if (m_p_qp_mgr) {
 		// 'down' the active QP/CQ
 		m_p_qp_mgr->down();
+
+		// Allow all WEQ's to be FLUSHED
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test PASSed.
See http://hpc-master.lab.mtl.com:8080/job/accl-libvma-pr/99/ for details (Mellanox internal link).

OphirMunk:
Added delay will increase latency during ring migration. 
Would you like to add a configure option for enabling/disabling this delay?

rosenbaumalex:
@OphirMunk True, this delay might hurt. But RING destroy and bring-up takes time anyway, so this delay is not as critical as you would expect.
Still I agree we should find a better way to make sure all FLUSHed cqe's arrived.
I'll take another look.

PR #260, opened 2016-12-18
issue: 924683 fix source address issue when using multicast connection with IP_MULTICAST_IF option

ashanably:
Signed-off-by: ashanably <ashanably@asaltech.com>


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/accl-libvma-pr/686/ for details (Mellanox internal link).

mellanox-github:
Test PASSed.
See http://hpc-master.lab.mtl.com:8080/job/accl-libvma-pr/676/ for details (Mellanox internal link).

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/accl-libvma-pr/696/ for details (Mellanox internal link).

src/vma/proto/dst_entry_udp_mc.cpp
```diff
@@ -81,6 +81,24 @@ bool dst_entry_udp_mc::conf_l2_hdr_and_snd_wqe_ib()
 	return ret_val;
 }
 
+void dst_entry_udp_mc::set_src_addr()
+{
+	m_pkt_src_ip = INADDR_ANY;
+	
+	if (m_bound_ip) {
+		m_pkt_src_ip = m_bound_ip;
```
OphirMunk: Following bind() m_pkt_src_ip will be updated. Is it also applicable to UC case? If not - please justify

ashanably: Described case is already handled in UC since in dst_entry::set_src_addr() we set m_pkt_src_ip to m_route_src_ip which will include m_bound_ip. However, in MC we don't update m_route_src_ip since we have different implementation for resolve_net_dev in dst_entry_udp_mc class and as result we used m_bound_ip directly.

OphirMunk: If I execute bind() after sending the first packet:
Source IP for MCAST packet will be different than source IP for Unicast packet.
If correct - please justify this difference

ashanably: If that case is allowed, then set_bound_addr will call set_state(false), as result resolve_net_dev will called again and assign m_route_src_ip one more time (currently we are not supporting multiple bind since kernel doesn't allow it).

src/vma/proto/dst_entry_udp_mc.cpp
```diff
@@ -81,6 +81,24 @@ bool dst_entry_udp_mc::conf_l2_hdr_and_snd_wqe_ib()
...
+	}
+	else if (m_mc_tx_if_ip.get_in_addr() && !m_mc_tx_if_ip.is_mc()) {
```
OphirMunk: Please explain why mc_tx_if is less prioritized compared to m_bound_ip for MCAST flows.

PR #311, opened 2017-02-27
issue: 972524 Optimize data-path by removing unnecessary if statements

rafiw:
Bring back if in datapath only before calling OS functions.
When application start it can call VMA before vma itself
starts, that why those checks are needed.

### DIFF
--- a/src/vma/sock/sock-redirect.cpp
+++ b/src/vma/sock/sock-redirect.cpp
@@ -341,6 +341,9 @@ int vma_recvfrom_zcopy(int __fd, void *__buf, size_t __nbytes, int *__flags,
 		return p_socket_object->rx(RX_RECVFROM, piov, 1, __flags, __from, __fromlen);
 
 	}
+	BULLSEYE_EXCLUDE_BLOCK_START
+	if (!orig_os_api.recvfrom) get_orig_funcs();
+	BULLSEYE_EXCLUDE_BLOCK_END
 	return orig_os_api.recvfrom(__fd, __buf, __nbytes, *__flags, __from, __fromlen);
 }
 
@@ -576,6 +579,10 @@ int socket_internal(int __domain, int __type, int __protocol, bool check_offload
 
 	dbg_check_if_need_to_send_mcpkt();
 
+	BULLSEYE_EXCLUDE_BLOCK_START
+	if (!orig_os_api.socket) get_orig_funcs();
+	BULLSEYE_EXCLUDE_BLOCK_END
+
 	int fd = orig_os_api.socket(__domain, __type, __protocol);
 	vlog_printf(VLOG_DEBUG, "ENTER: %s(domain=%s(%d), type=%s(%d), protocol=%d) = %d\n",__func__, socket_get_domain_str(__domain), __domain, socket_get_type_str(__type), __type, __protocol, fd);
 	if (fd < 0) {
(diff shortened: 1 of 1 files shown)


mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/accl-libvma-pr/1208/ for details (Mellanox internal link).

mellanox-github:
Test PASSed.
See http://hpc-master.lab.mtl.com:8080/job/accl-libvma-pr/853/ for details (Mellanox internal link).

src/vma/sock/sock-redirect.cpp
```diff
@@ -724,6 +763,10 @@ int bind(int __fd, const struct sockaddr *__addr, socklen_t __addrlen)
 extern "C"
 int connect(int __fd, const struct sockaddr *__to, socklen_t __tolen)
 {
+	BULLSEYE_EXCLUDE_BLOCK_START
```
OphirMunk: @rafiw Here you are calling get_orig_funcs() at function startup. Which is OK since connect() is not on the fast path. However, please look below at setsockopt()...

src/vma/sock/sock-redirect.cpp
```diff
@@ -776,6 +819,9 @@ int setsockopt(int __fd, int __level, int __optname,
 		ret = p_socket_object->setsockopt(__level, __optname, __optval, __optlen);
 	}
 	else {
+		BULLSEYE_EXCLUDE_BLOCK_START
```
OphirMunk: Here you are delaying the call to get_orig_funcs() just before needing orig_os_api.setsockopt, which is also OK.
However need to keep the same code style (either calling at startup or just before needing which is the preferred style in my view).
Please rewrite keeping the same code style

rafiw: i preferred calling just before it's needed
the reason i called in accept and connect at the the beginning of the function is that there is more then one flow calling the os function e.g. in connect
https://github.com/Mellanox/libvma/pull/311/files#diff-c49e7dd4d795b90a21b8c1708ff6d7e8L739
https://github.com/Mellanox/libvma/pull/311/files#diff-c49e7dd4d795b90a21b8c1708ff6d7e8L747
i wanted to keep it simple
the other option is to rewrite the connect logic to something like this (didn't test it and has an issue but that's the spirit)

```
if (__to && __to->sa_family == AF_INET && p_socket_object) {
	ret = p_socket_object->connect(__to, __tolen);
	if (p_socket_object->isPassthrough()) {
		handle_close(__fd, false, true);
		if (ret == 0) {
			srdr_logdbg_exit("failed (errno=%d %m)", errno);
			return ret;
		}

	}
}
if (p_socket_object) {
	p_socket_object->setPassthrough();
}
BULLSEYE_EXCLUDE_BLOCK_START
if (!orig_os_api.connect) get_orig_funcs();
BULLSEYE_EXCLUDE_BLOCK_END
ret = orig_os_api.connect(__fd, __to, __tolen);

if (ret >= 0)
	srdr_logdbg_exit("returned with %d", ret);
else
	srdr_logdbg_exit("failed (errno=%d %m)", errno);
return ret;
```

OphirMunk: Explanation understood. Let's remain with your original code. No need to change.

OphirMunk:
Merged #311

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/accl-libvma-pr/1234/ for details (Mellanox internal link).

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/accl-libvma-pr/1257/ for details (Mellanox internal link).

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/accl-libvma-pr/1273/ for details (Mellanox internal link).

mellanox-github:
Test PASSed.
See http://bgate.mellanox.com/jenkins/job/accl-libvma-pr/1292/ for details (Mellanox internal link).

PR #463, opened 2017-09-12
issue: 1134718 Fix lack of interface reinitialization (Tx PRM)

DanielLibenson:
PRM Tx post send queue initialization function move from constructor
to qp::up(), to support interface bring up after being down.


liranoz12:
Merged #463

PR #477, opened 2017-10-01
issue: 1150011 Fix ioctl() FIONREAD return value while no data is ava…

liranoz12:
…ilable

During ioctl() with FIONREAD call, In order to verify if there is data
available for reading, VMA calls non blocking rx_wait().
When no data was available, we would return an error (EAGAIN).


liranoz12:
Merged #477

PR #478, opened 2017-10-01
issue: 1154650 Fix IP fragmentation threshold

liranoz12:
The decision on the maximum size to be fragmented should be based on
the MTU, IP header (20 bytes) and UDP header (8 bytes) without any
alignment.
When fragmentation is performed, the size of each fragment should be
aligned to 8 bytes because of the limits of the offset field.

### DIFF
--- a/src/vma/proto/dst_entry.h
+++ b/src/vma/proto/dst_entry.h
@@ -134,8 +134,8 @@ protected:
 	vma_ibv_send_wr* 	m_p_send_wqe;
 	uint32_t 		m_max_inline;
 	ring_user_id_t		m_id;
-	size_t			m_max_ip_payload_size;
-	size_t			m_max_udp_payload_size;
+	uint16_t		m_max_ip_payload_size;
+	uint16_t		m_max_udp_payload_size;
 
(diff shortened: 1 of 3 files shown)


src/vma/proto/dst_entry.h
```diff
@@ -133,8 +133,9 @@ class dst_entry : public cache_observer, public tostr, public neigh_observer
 
 	vma_ibv_send_wr* 	m_p_send_wqe;
 	uint32_t 		m_max_inline;
-	ring_user_id_t 	m_id;
+	ring_user_id_t		m_id;
 	size_t			m_max_ip_payload_size;
```
rafiw: consider using uint_16 for both parameters

liranoz12: This will cause me to change type in several places. for example :

proto/dst_entry_udp.cpp:215:85: error: no matching function for call to 'min(uint16_t&, size_t)'
   size_t sz_ip_frag = min(m_max_ip_payload_size, (sz_udp_payload - n_ip_frag_offset));

I dont think it worth it..

liranoz12:
Merged #478

PR #479, opened 2017-10-02
issue: 1148076 Remove volatile when handling CQE in MP_RQ

rafiw:


rafiw:
@olgk please take a look thanks

PR #480, opened 2017-10-03
issue: 1160823 fix crash when reg_mr fails

rafiw:

### DIFF
--- a/src/vma/dev/allocator.cpp
+++ b/src/vma/dev/allocator.cpp
@@ -186,7 +186,6 @@ bool vma_allocator::register_memory(size_t size, ib_ctx_handler *p_ib_ctx_h,
 {
 	bool failed = false;
 	if (p_ib_ctx_h) {
-		m_mr_list_len = 1;
 		m_mr_list = new ibv_mr*[1];
 		m_mr_list[0] = p_ib_ctx_h->mem_reg(m_data_block, size, access);
 		BULLSEYE_EXCLUDE_BLOCK_START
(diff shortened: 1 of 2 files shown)


src/vma/dev/allocator.cpp
```diff
@@ -234,6 +234,10 @@ bool vma_allocator::register_memory(size_t size, ib_ctx_handler *p_ib_ctx_h,
 vma_allocator::~vma_allocator() {
 	// Unregister memory
 	for (size_t i = 0; i < m_mr_list_len; ++i) {
+		if (!m_mr_list[i]) {
+			__log_info_err("ibv_mr is null, list length is %zd", m_mr_list_len);
+			continue;
+		}
```
liranoz12: It does not make sense to have a NULL in mr_list.
A better approach would be not to add NULL element to the list when the registration failed.
Another weak spot :
`uint32_t vma_allocator::find_lkey_by_ib_ctx(ib_ctx_handler *p_ib_ctx_h) const
{
        ibv_device* dev = p_ib_ctx_h->get_ibv_device();
        for (size_t i = 0; i < m_mr_list_len; ++i) {
                if (dev == m_mr_list[i]->context->device) { <---- crash
                        return m_mr_list[i]->lkey;
                }
        }
        return 0;
}
`
Let's talk about it on Sunday.

rafiw: this flow will never happen due to the fact that reg_mr fails so VMA exits.
but i handled the cause of the bug (bad array size setting) in this new commit

liranoz12:
Merged #480

PR #488, opened 2017-10-25
issue: 1175096 VMA failed to detect FILLER in MP_RQ

rafiw:
VMA assumed that in filler messages the IP CSUM bit is off
this assumption is not correct.


rafiw:
fixing cases with fillers is more complicated then this commit alone.
so currently i'm closing this PR until we decide we want to support fillers (msg size == odd number of stride size, e.g. stirde =512 msg - 1400.)

PR #491, opened 2017-11-01
issue: 1181379 Fix a TCP zero-wnd probe when there is data in-flight

DanielLibenson:
TCP zero-window probe message should not be sent when there
is data in-flight(unacked packets, pcb->unacked != NULL).


rafiw:
bot:retest

liranoz12:
bot:retest

liranoz12:
Merged #491

PR #494, opened 2017-11-05
issue: 1185372 Traffic is not offloaded for high VMA_RX_WRE values

liranoz12:
As part of VMA's initialization process, we are trying to open QP on
each ETH device (verify_eth_qp_creation()) which fails for high
VMA_RX_WRE values.
This happens because VMA is trying to open QP using high value of
VMA_RX_WRE which is not supported.
A better way to handle this is to use a default value of VMA_RX_WRE at
verify_eth_creation().


liranoz12:
Merged #494

PR #499, opened 2017-11-09
issue: 1175657 Fix ring statistics are not updated when flowtag is ena…

liranoz12:
…bled


liranoz12:
Merged #499

PR #500, opened 2017-11-12
issue: 1190561 detect change route dynamiclly

rafiw:
This commit adds support for detection of routing changes
detected by libnl.
Perior to this commmit:
1. There was no seperatin between new route and change rotue
2. Route invalidatioin was not discovered by the dst_entry.
This commit fixes both issues.

note: route delete is still not detected in libvma, so libvma
      will search the route table from the end. This will
      solve some of the issues (e.g. del route and adding new one),
      but this will not solve cases of just deleting routes and
      fallback to default\existing rule.


rafiw:
this could be problematic in case socket need to move between rings or devices.
effort is to big for now.

PR #507, opened 2017-11-20
issue: 1213478 Fix flow is attached to incorrect ring

liranoz12:
Although VMA creates net_device_val for each IP, while binding TCP
socket to a secondary ip, the established socket is attached to the
ring of the primary IP.


liranoz12:
From VMA 8.6, Net devices objects are created per interface index and not per ip.

PR #509, opened 2017-11-21
issue:1214066 add support for extended table id

rafiw:
This commit adds support for parsing route and rule
extended table_id


rafiw:
https://www.mail-archive.com/netdev@vger.kernel.org/msg15604.html
modern Linux kernels support for RTA_TABLE rtnetlink attribute.
For compatibility with previous kernels rtm_table field in struct rtmsg is 
used only for values less than 256."
extended table means table id can be uint32_t this is a new mode

liranoz12:
bot:retest

PR #512, opened 2017-11-22
issue: 1213984 use dst_enty to find the route MTU

rafiw:
VMA used only the IP destination address to find the MTU route.
This caused VMA not to consider source based rules.
This commit removes this logic and uses the dst_entry
in the socket to get the route and the MTU.

* 1c981ad issue: 1213984 use dst_enty to find the route MTU


liranoz12:
Merged #512

PR #521, opened 2017-12-15
Pr/issue 1228170

serglelv:
Enable RST for incoming packet that has the ACK bit set but not the SYN bit set


serglelv:
bot:retest

PR #522, opened 2017-12-15
issue: 1228170 Fix enable outgoing RST

serglelv:
Enable outgoing RST packet  for the incoming packet that has the ACK bit set but not the SYN bit set.


serglelv:
Please review. @liranoz12

PR #524, opened 2017-12-19
issue: 1235797 Not respond to incoming SYN-FIN in the listen state

serglelv:
TCP segment with the SYN-FIN-bits set in the listen state should be ignored


serglelv:
@liranoz12 Please review

liranoz12:
@serglelv,
As we discussed, Please move the new "if" statement to the upper section of the function (same as #523)
Thanks,

liranoz12:
Merged #524

PR #537, opened 2018-01-03
issue: 1248705 add extra api to get ring TX fd from socket

rafiw:

### DIFF
--- a/src/vma/proto/dst_entry.h
+++ b/src/vma/proto/dst_entry.h
@@ -98,6 +98,7 @@ public:
 	void		return_buffers_pool();
 	int		get_route_mtu();
 	inline header*	get_network_header() { return &m_header;}
+	inline ring*	get_ring() { return m_p_ring;}
 protected:
 	ip_address 		m_dst_ip;
 	uint16_t 		m_dst_port;
--- a/src/vma/sock/sockinfo_udp.cpp
+++ b/src/vma/sock/sockinfo_udp.cpp
@@ -541,6 +541,7 @@ int sockinfo_udp::connect(const struct sockaddr *__to, socklen_t __tolen)
 	// Dissolve the current connection setting if it's not AF_INET
 	// (this also support the default dissolve by AF_UNSPEC)
 	if (connect_to.get_sa_family() == AF_INET) {
+		m_connected.set_sa_family(AF_INET);
 		m_connected.set_in_addr(INADDR_ANY);
 		m_p_socket_stats->connected_ip = m_connected.get_in_addr();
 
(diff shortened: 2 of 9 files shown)


src/vma/sock/sockinfo_udp.cpp
```diff
@@ -2760,6 +2761,42 @@ size_t sockinfo_udp::handle_msg_trunc(size_t total_rx, size_t payload_size, int
 	return total_rx;
 }
 
+int sockinfo_udp::get_socket_tx_ring_fd(struct sockaddr *to, socklen_t tolen)
+{
+	NOT_IN_USE(tolen);
+	si_udp_logfunc("get_socket_tx_ring_fd fd %d to %p tolen %d", m_fd, to ,tolen);
+
```
liranoz12: Why does "tolen" parameter exist if we don't use it at all ?

src/vma/sock/sockinfo_udp.cpp
```diff
@@ -2760,6 +2761,42 @@ size_t sockinfo_udp::handle_msg_trunc(size_t total_rx, size_t payload_size, int
...
+		errno = EINVAL;
+		return -1;
+	}
+	sock_addr dst(to);
+	ring *ring = NULL;
+
+	if (m_p_connected_dst_entry && m_connected == dst) {
+		ring = m_p_connected_dst_entry->get_ring();
+	} else {
+		dst_entry_map_t::iterator it = m_dst_entry_map.begin();
```
liranoz12: Please consider using "find" instead of for loop as here:
https://github.com/Mellanox/libvma/blob/master/src/vma/sock/sockinfo_udp.cpp#L1801

src/vma/sock/sockinfo_udp.cpp
```diff
@@ -2760,6 +2761,42 @@ size_t sockinfo_udp::handle_msg_trunc(size_t total_rx, size_t payload_size, int
...
+			if (it->first == dst) {
+				ring = it->second->get_ring();
+				break;
+			}
+		}
+	}
+	if (!ring) {
+		si_udp_logdbg("could not find TX ring for fd %d addr %s",
+				m_fd, dst.to_str());
+		errno = ENODATA;
```
liranoz12: Maybe ENOTCONN ?

src/vma/proto/dst_entry.h
```diff
@@ -98,6 +98,7 @@ class dst_entry : public cache_observer, public tostr, public neigh_observer
 	void		return_buffers_pool();
 	int		get_route_mtu();
 	inline header*	get_network_header() { return &m_header;}
+	inline ring*	get_ring() { return m_p_ring;}
```
liranoz12: Why to expose the ring if we only want to the tx channel fd ?

src/vma/sock/sockinfo_udp.cpp
rafiw: not good since there can be a connection but it's not offloaded

src/vma/proto/dst_entry.h
rafiw: more generic for future needs.
ring can be null

src/vma/sock/sockinfo_udp.cpp
rafiw: always when passing address it's len is passed, this keeps some alignment with other socket based calls (connect bind).
i can add (maybe should) add a test to verify the size

rafiw:
issue: 1248705 add extra api to get the TX ring fd from the socket

This commit adds support to retrieve the FD of the TX ring used by the socket
to send the packets. This FD can be used to create eth_direct ring and do PRM
data path in the upper application

PR #540, opened 2018-01-04
issue: 1247358 Fix crash when VMA could not resolve neighbour

liranoz12:


MohammadQurt:
I think the "m_val is not valid" check is an important part of fixing the issue. 
Thanks

PR #543, opened 2018-01-09
issue: 1261518 remove unneeded lock

rafiw:
Use the bufferPool lock instead of the stats lock


rafiw:
not important

PR #554, opened 2018-01-17
issue: 1234612 Fix SYN_RCVD via SYN received

serglelv:


serglelv:
@liranoz12 Hi Liran, please review.

serglelv:
bot:retest

src/vma/lwip/tcp.c
```diff
@@ -82,6 +82,13 @@ void register_tcp_seg_free(tcp_seg_free_fn fn)
 }
 #endif
 
+tcp_set_error_staus_fn external_tcp_set_error_status;
+
+void register_tcp_set_error_status(tcp_set_error_staus_fn fn)
+{
+	external_tcp_set_error_status = fn;
+}
+
```
liranoz12: Consider moving m_error_status into the pcb struct.
This way you will not need to add this callback.

src/vma/lwip/tcp_out.c
```diff
@@ -1121,6 +1121,11 @@ tcp_output(struct tcp_pcb *pcb)
        #if TCP_OVERSIZE_DBGCHECK
          seg->oversize_left = 0;
        #endif /* TCP_OVERSIZE_DBGCHECK */
+       if ((pcb->in_flag_hndlg & TCP_SYN) && (get_tcp_state(pcb) & SYN_RCVD)) {
+         pcb->in_flag_hndlg = 0;
+         pcb->rcv_nxt = 1;
+         seg->tcphdr->seqno = htonl(htonl(seg->tcphdr->seqno)-1);
+       }
```
liranoz12: This will probably affect the fast path, can we avoid this addition ?

src/vma/sock/sockinfo_tcp.cpp
```diff
@@ -3719,6 +3720,31 @@ int sockinfo_tcp::getsockopt_offload(int __level, int __optname, void *__optval,
...
+				ti->tcpi_rcv_mss = m_pcb.mss;
+				ti->tcpi_advmss = m_pcb.advtsd_mss;
+				ti->tcpi_snd_cwnd = m_pcb.cwnd;
+				ti->tcpi_snd_ssthresh = m_pcb.ssthresh;
+				ti->tcpi_unacked = m_pcb.unacked->seqno;
+				ret = 0;
+			} else {
+				errno = EINVAL;
+			}
+			break;
```
liranoz12: Is this related to the fix? if not we need to split it into two commits.

src/vma/lwip/tcp_in.c
```diff
@@ -599,6 +599,16 @@ tcp_process(struct tcp_pcb *pcb, tcp_in_data* in_data)
...
+    /* received SYN? SYN_RCVD via SYN */
+    else if (in_data->flags & TCP_SYN) {
+      err = tcp_enqueue_flags(pcb, TCP_SYN);
+      if (err == ERR_OK) {
+        /* Changed the pcbs state now */
+        set_tcp_state(pcb, SYN_RCVD);
+        pcb->in_flag_hndlg |= TCP_SYN;
+        TCP_SET_ERROR_STATUS(pcb, 0);
+      }
+    }
```
liranoz12: 1. Is this addition supposed to handle this case ?
https://tools.ietf.org/html/rfc793#section-3.4 
The "three-way handshake" is the procedure used to establish a
  connection.  This procedure normally is initiated by one TCP and
  responded to by another TCP.  The procedure also works if two TCP
  simultaneously initiate the procedure.  When simultaneous attempt
  occurs, each TCP receives a "SYN" segment which carries no
  acknowledgment after it has sent a "SYN". 

2. tcp_enqueue_flags is used to enqueue TCP options for transmission but i dont see any tcp_out() call here ?
Where we send the data ?

3. set_tcp_state(pcb, SYN_RCVD) is called only from tcp_listen_input() while with this change it can also be called from tcp_process during connect() - is this valid ?

src/vma/sock/sockinfo_tcp.cpp
serglelv: This code related to the fix but only used in PD.  I agree this code should be removed.

src/vma/lwip/tcp_out.c
serglelv: This functionality probably can be moved to the slowtmr/fasttmr? I will review it.

src/vma/lwip/tcp_in.c
serglelv: 1. It seems yes. 2. The fast path is used. Review -wip. 3. tcp connect during connect  review - wip.

src/vma/lwip/tcp.c
serglelv: I agree. The code will be updated.

PR #556, opened 2018-01-18
issue: 1272179 Arrange Ring statistics

liranoz12:
Statistics regarding ring interrupts, moderations and retransmissions
    will be published only when the value is not zero.

    Signed-off-by: Liran Oz <lirano@mellanox.com


liranoz12:
bot:retest

PR #559, opened 2018-01-18
issue: 1262198 Fix epoll_wait returns fd event twice in single call

liranoz12:


FerasHamad:
PR 559 Fixes the issue
Approved

PR #564, opened 2018-01-25
issue: 1262560 Fix zero-window ack probe

serglelv:
TCP keepalive should be sent if zero-window ack probe was received


serglelv:
@liranoz12 Hi Liran, please review.

liranoz12:
duplicated by PR 590

PR #589, opened 2018-03-12
issue: 1330652 get rate limit as KBit per sec in new interface

rafiw:
When setting vma_rate_limit_t expect rate as kbit instead
of bytes. This is needed since uint32 is not big enough for
big values of bytes per second

### DIFF
--- a/src/vma/sock/sockinfo.h
+++ b/src/vma/sock/sockinfo.h
@@ -57,7 +57,7 @@
 #define BASE_SOCKINFO_H
 
 #define SI_RX_EPFD_EVENT_MAX		16
-#define BYTE_TO_KB(BYTEVALUE)		(((BYTEVALUE) * 8) / 1000)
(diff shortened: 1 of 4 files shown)


rafiw:
@ilansmith please review

src/vma/sock/sockinfo.h
```diff
@@ -57,7 +57,7 @@
 #define BASE_SOCKINFO_H
 
 #define SI_RX_EPFD_EVENT_MAX		16
-#define BYTE_TO_KB(BYTEVALUE)		(((BYTEVALUE) * 8) / 1000)
+#define BYTE_TO_KB(BYTEVALUE)		((uint64_t)((uint64_t)((uint64_t)BYTEVALUE) * 8) / 1000)
```
ilansmith: I changed to ((byte_value) / 125) and for the row bellow ((kbit_value) * 125).

ilansmith:
Replaced by #591

PR #590, opened 2018-03-13
issue: 1262560 Fix zero-window ack probe

liranoz12:
TCP zero-window-probe should be sent as long as the window is full.


liranoz12:
@serglelv - Please review and confirm with PD.
@mtami - Please verify that the PR solves the reported issue.

This PR duplicates #564.

PR #591, opened 2018-03-14
issue: 1330652 Use kbps value for setting enhanced rate limit

ilansmith:
The value for setting socket option SO_MAX_PACING_RATE is provided in
bytes.
For the enhanced version of this socket option (where we set rate,
burst size and packet size) the rate is provided in Kbps to prevent
an unnecessary double conversion and loss of accuracy.

### DIFF
--- a/src/vma/sock/sockinfo.cpp
+++ b/src/vma/sock/sockinfo.cpp
@@ -283,7 +283,6 @@ int sockinfo::getsockopt(int __level, int __optname, void *__optval, socklen_t *
 		case SO_MAX_PACING_RATE:
 			if (*__optlen == sizeof(struct vma_rate_limit_t)) {
 				*(struct vma_rate_limit_t*)__optval = m_so_ratelimit;
-				(*(struct vma_rate_limit_t*)__optval).rate = m_so_ratelimit.rate;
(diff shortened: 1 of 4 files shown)


ilansmith:
Replaces #589.
Was coded by @rafiw and reviewed with some corrections made by me.

src/vma/sock/sockinfo.cpp
```diff
@@ -281,15 +281,19 @@ int sockinfo::getsockopt(int __level, int __optname, void *__optval, socklen_t *
 #endif // DEFINED_SOCKETXTREME
 
 		case SO_MAX_PACING_RATE:
-			if (*__optlen >= sizeof(struct vma_rate_limit_t)) {
+			if (*__optlen == sizeof(struct vma_rate_limit_t)) {
 				*(struct vma_rate_limit_t*)__optval = m_so_ratelimit;
-				(*(struct vma_rate_limit_t*)__optval).rate = KB_TO_BYTE(m_so_ratelimit.rate);
+				(*(struct vma_rate_limit_t*)__optval).rate = m_so_ratelimit.rate;
```
rafiw: this line can be removed

PR #627, opened 2018-05-23
issue: 1387232 Add vmad configurable notify dir

rafiw:
By dafault VMA writes the fd output files to the folder
/tmp/vma/, and daemon reds them from there.
Now user can change this folder location,
VMA should be run with VMA_VMAD_NOTIFY_DIR and
vmad should be run with --notify-dir.
default folder remains /tmp/vma/


liranoz12:
+1

PR #657, opened 2018-06-14
issue: 1424602 Fix crash while valgrind is loaded without SRIOV

liranoz12:
Microsoft Hyper detection does not work without SRIOV while valgrind
is loaded (see third banner).
This will cause VMA to be load with contig pages - which will cause
a core dump.


rafiw:
Crush or crash?

PR #659, opened 2018-06-17
issue: 1427833 Update ib_ctx collection table by parameter

liranoz12:
There is no need to iterate over all existing IB devices because
all information is received by netlink.

Signed-off-by: Liran Oz <lirano@mellanox.com


liranoz12:
@igor-ivanov please review

PR #700, opened 2018-08-09
issue: 1498320 Set RDMAV_ALLOW_DISASSOC_DESTROY env for upstream

liranoz12:
RDMAV_ALLOW_DISASSOC_DESTROY tells ibv_destroy functions we
want to get success errno value in case of calling them
when the device was removed.


liranoz12:
bot:retest

PR #710, opened 2018-08-29
issue: 1087459 Fix VMA_EXCEPTION_HANDLING=0/1 generates core

liranoz12:
un-offloading the socket from the socket context is not allowed.


liranoz12:
bot:retest

PR #735, opened 2018-10-28
issue: 1438405 Update ip header after dst entry creation

rafiw:


src/vma/proto/dst_entry.cpp
```diff
@@ -376,7 +376,9 @@ bool dst_entry::conf_l2_hdr_and_snd_wqe_eth()
 		if (src && dst) {
 		BULLSEYE_EXCLUDE_BLOCK_END
 			if (netdevice_eth->get_vlan()) { //vlan interface
-				uint16_t vlan_tci = (m_pcp << 12) | netdevice_eth->get_vlan();
+				uint16_t vlan_tci =
+						((uint16_t)m_pcp << NET_ETH_VLAN_PCP_OFFSET) |
```
liranoz12: pcp value was 12 before this change, are we sure about it ?

src/vma/proto/header.h
```diff
@@ -54,7 +54,7 @@
 #define NET_IB_IP_ALIGN_SZ		16
 #define NET_ETH_IP_ALIGN_SZ		 6
 #define NET_ETH_VLAN_IP_ALIGN_SZ	 2
-
+#define NET_ETH_VLAN_PCP_OFFSET	13
```
liranoz12: Can we use kernel definition instead ?

src/vma/proto/dst_entry.h
```diff
@@ -151,8 +153,8 @@ class dst_entry : public cache_observer, public tostr, public neigh_observer
 	virtual uint8_t 	get_protocol_type() const = 0;
 	virtual bool 		get_net_dev_val();
 	virtual uint32_t 	get_inline_sge_num() = 0;
-	virtual ibv_sge*	get_sge_lst_4_inline_send() = 0;
 	virtual ibv_sge*	get_sge_lst_4_not_inline_send() = 0;
+	ibv_sge*		get_sge_lst_4_inline_send() { return m_sge; };
```
liranoz12: Please update commit message with issue id

src/vma/proto/header.cpp
```diff
@@ -39,25 +39,29 @@ void header::init()
 	memset(&m_header, 0, sizeof(m_header));
 	m_ip_header_len = 0;
 	m_transport_header_len = 0;
-	m_udp_header_len = 0;
 	m_total_hdr_len = 0;
 	m_aligned_l2_l3_len = 40;
+	m_is_valn_enabled = false;
```
liranoz12: should be is_vlan_enabled ?

src/vma/sock/sockinfo_tcp.cpp
```diff
@@ -3443,6 +3443,17 @@ int sockinfo_tcp::setsockopt(int __level, int __optname,
 				val = *(int *)__optval;
 				val &= ~INET_ECN_MASK;
 				m_pcb.tos |= val & INET_ECN_MASK;
+				header_tos_updater du(m_pcb.tos);
+				update_header_field(&du);
+				// according to the linux
```
liranoz12: Please add more detailed explanation

src/vma/sock/sockinfo_udp.cpp
```diff
@@ -1216,6 +1218,16 @@ int sockinfo_udp::setsockopt(int __level, int __optname, __const void *__optval,
 			case IP_TOS:
 				if (__optlen <= sizeof(int)) {
 					m_tos =(uint8_t) *(int *)__optval;
+					header_tos_updater du(m_pcp);
```
liranoz12: Can we move IP_TOS to sockinfo and avoid the duplication  ?

src/vma/sock/sockinfo_udp.cpp
```diff
@@ -2817,18 +2829,18 @@ bool sockinfo_udp::prepare_to_close(bool process_shutdown) {
...
+	for (; dst_entry_iter != m_dst_entry_map.end(); dst_entry_iter++) {
 		if (!IN_MULTICAST_N(dst_entry_iter->second->get_dst_addr())) {
-			dst_entry_iter->second->set_ip_ttl(m_n_uc_ttl);
+			updater->update_field(*dst_entry_iter->second);
 		}
 	}
 
-	if (m_p_connected_dst_entry && !IN_MULTICAST_N(m_p_connected_dst_entry->get_dst_addr())) {
-		m_p_connected_dst_entry->set_ip_ttl(m_n_uc_ttl);
+	if (m_p_connected_dst_entry &&
```
liranoz12: Do we want to ignore it for MC ?

src/vma/proto/dst_entry.cpp
rafiw: yes i doubled check and compared to the kernel

src/vma/proto/header.h
rafiw: there isn't one

rafiw: there isn't one

src/vma/proto/header.cpp
rafiw: done

src/vma/sock/sockinfo_tcp.cpp
rafiw: added

src/vma/sock/sockinfo_udp.cpp
rafiw: there is a difference in behavior

liranoz12:
Duplicated by #741

PR #741, opened 2018-11-01
issue: 1438405 Allow dynamic updating header bits after ring creation

rafiw:


src/vma/sock/sockinfo_udp.cpp
```diff
@@ -939,13 +941,20 @@ int sockinfo_udp::setsockopt(int __level, int __optname, __const void *__optval,
...
-					if (__optval && (n_mc_ttl >= 0 && n_mc_ttl <= 255)) {
+					else {
+						break;
+					}
+					if (n_mc_ttl == -1) {
+						n_mc_ttl = 1;
+					}
+					if (n_mc_ttl >= 0 && n_mc_ttl <= 255) {
 						m_n_mc_ttl = n_mc_ttl;
+						header_tos_updater du(m_n_mc_ttl);
```
liranoz12: header_tos_updater type is wrong...
And i think we should add another type for MC.. udp socket might have diffent mc/uc ttl's ..

PR #792, opened 2019-01-03
issue: 1605819 Add ring allocation logic control for all socket types

Awik84:
Allow user to modify the socket's ring allocation logic
via setsockopt.

Signed-off-by: Ariel Levkovich <lariel@mellanox.com>

### DIFF
--- a/src/stats/stats_printer.cpp
+++ b/src/stats/stats_printer.cpp
@@ -129,6 +129,15 @@ void print_full_stats(socket_stats_t* p_si_stats, mc_grp_info_t* p_mc_grp_info, 
 	}
 
 	//
+	// Ring Allocation Logic information
+	//
+	//
+	if (p_si_stats->ring_alloc_logic_rx == RING_LOGIC_PER_USER_ID)
+		fprintf(filename, "- RX: Ring User ID = %lu\n", p_si_stats->ring_user_id_rx);
+	if (p_si_stats->ring_alloc_logic_tx == RING_LOGIC_PER_USER_ID)
+		fprintf(filename, "- TX: Ring User ID = %lu\n", p_si_stats->ring_user_id_tx);
+
+	//
 	// Socket statistics
 	//
 	if (p_si_stats->counters.n_tx_sent_byte_count || p_si_stats->counters.n_tx_sent_pkt_count || p_si_stats->counters.n_tx_drops || p_si_stats->counters.n_tx_errors)
@@ -187,16 +196,6 @@ void print_full_stats(socket_stats_t* p_si_stats, mc_grp_info_t* p_mc_grp_info, 
 	if (b_any_activiy == false) {
 		fprintf(filename, "Rx and Tx where not active\n");
 	}
-
-	//
-	// Ring Allocation Logic information
-	//
-	//
-	if (p_si_stats->ring_alloc_logic_rx == RING_LOGIC_PER_USER_ID)
-		fprintf(filename, "- RX: Ring User ID = %lu\n", p_si_stats->ring_user_id_rx);
-	if (p_si_stats->ring_alloc_logic_tx == RING_LOGIC_PER_USER_ID)
-		fprintf(filename, "- TX: Ring user ID = %lu\n", p_si_stats->ring_user_id_tx);
-
 }
 
 // Print statistics headers for all sockets - used in case view mode is e_netstat_like
--- a/src/stats/stats_printer.cpp
+++ b/src/stats/stats_printer.cpp
@@ -106,15 +106,6 @@ void print_full_stats(socket_stats_t* p_si_stats, mc_grp_info_t* p_mc_grp_info, 
 	fprintf(filename, "\n");
 
 	//
-	// Ring Allocation Logic information
-	//
-	//
-	if (p_si_stats->ring_alloc_logic_rx == RING_LOGIC_PER_USER_ID)
-		fprintf(filename, "- RX: Ring User ID = %lu\n", p_si_stats->ring_user_id_rx);
-	if (p_si_stats->ring_alloc_logic_tx == RING_LOGIC_PER_USER_ID)
-		fprintf(filename, "- TX: Ring user ID = %lu\n", p_si_stats->ring_user_id_tx);
-
-	//
 	// Bounded + Connected information
 	//
 	if (p_si_stats->bound_if || p_si_stats->bound_port) {
@@ -196,6 +187,16 @@ void print_full_stats(socket_stats_t* p_si_stats, mc_grp_info_t* p_mc_grp_info, 
 	if (b_any_activiy == false) {
 		fprintf(filename, "Rx and Tx where not active\n");
 	}
+
+	//
+	// Ring Allocation Logic information
+	//
+	//
+	if (p_si_stats->ring_alloc_logic_rx == RING_LOGIC_PER_USER_ID)
+		fprintf(filename, "- RX: Ring User ID = %lu\n", p_si_stats->ring_user_id_rx);
+	if (p_si_stats->ring_alloc_logic_tx == RING_LOGIC_PER_USER_ID)
+		fprintf(filename, "- TX: Ring user ID = %lu\n", p_si_stats->ring_user_id_tx);
+
 }
 
 // Print statistics headers for all sockets - used in case view mode is e_netstat_like
--- a/src/stats/stats_printer.cpp
+++ b/src/stats/stats_printer.cpp
@@ -37,6 +37,8 @@
 #include "vma/util/utils.h"
 #include "vma/util/vma_stats.h"
 #include "vma/lwip/tcp.h"
+#include "vma/vma_extra.h"
+#include "vma/util/sys_vars.h"
 
 typedef enum {
 	e_K = 1024,
@@ -65,6 +67,22 @@ const char* to_str_socket_type_netstat_like(int type)
 	case SOCK_STREAM:	return "tcp";
 	case SOCK_DGRAM:	return "udp";
 	case SOCK_RAW:		return "raw";
+	default:
+		break;
+	}
+	return "???";
+}
+
+const char* to_str_ring_alloc_logic(ring_logic_t logic)
+{
+	switch (logic) {
+	case RING_LOGIC_PER_INTERFACE: return "RING_PER_IF";
+	case RING_LOGIC_PER_IP: return "RING_PER_IP";
+	case RING_LOGIC_PER_SOCKET: return "RING_PER_SOCKET";
+	case RING_LOGIC_PER_USER_ID: return "RING_PER_USER_ID";
+	case RING_LOGIC_PER_THREAD: return "RING_PER_THREAD";
+	case RING_LOGIC_PER_CORE: return "RING_PER_CORE";
+	case RING_LOGIC_PER_CORE_ATTACH_THREADS: return "RING_PER_CORE_ATTACH_THREADS";
 	default:
 		break;
 	}
@@ -102,6 +120,14 @@ void print_full_stats(socket_stats_t* p_si_stats, mc_grp_info_t* p_mc_grp_info, 
 		}
 	}
 	fprintf(filename, "\n");
+
+	//
+	// Ring Allocation Logic information
+	//
+	fprintf(filename, "- RX: ring alloc logic = %s, ring user ID = %lu\n",
+			ring_logic_str(p_si_stats->ring_alloc_logic_rx), p_si_stats->ring_user_id_rx);
+	fprintf(filename, "- TX: ring alloc logic = %s, ring user ID = %lu\n",
+			ring_logic_str(p_si_stats->ring_alloc_logic_tx), p_si_stats->ring_user_id_tx);
 
 	//
 	// Bounded + Connected information
--- a/src/stats/stats_printer.cpp
+++ b/src/stats/stats_printer.cpp
@@ -108,10 +108,11 @@ void print_full_stats(socket_stats_t* p_si_stats, mc_grp_info_t* p_mc_grp_info, 
 	//
 	// Ring Allocation Logic information
 	//
-	fprintf(filename, "- RX: ring alloc logic = %s, ring user ID = %lu\n",
-			ring_logic_str(p_si_stats->ring_alloc_logic_rx), p_si_stats->ring_user_id_rx);
-	fprintf(filename, "- TX: ring alloc logic = %s, ring user ID = %lu\n",
-			ring_logic_str(p_si_stats->ring_alloc_logic_tx), p_si_stats->ring_user_id_tx);
+	//
+	if (p_si_stats->ring_alloc_logic_rx == RING_LOGIC_PER_USER_ID)
+		fprintf(filename, "- RX: Ring User ID = %lu\n", p_si_stats->ring_user_id_rx);
+	if (p_si_stats->ring_alloc_logic_tx == RING_LOGIC_PER_USER_ID)
+		fprintf(filename, "- TX: Ring user ID = %lu\n", p_si_stats->ring_user_id_tx);
 
 	//
 	// Bounded + Connected information
--- a/src/stats/stats_printer.cpp
+++ b/src/stats/stats_printer.cpp
@@ -73,22 +73,6 @@ const char* to_str_socket_type_netstat_like(int type)
 	return "???";
 }
 
-const char* to_str_ring_alloc_logic(ring_logic_t logic)
-{
-	switch (logic) {
-	case RING_LOGIC_PER_INTERFACE: return "RING_PER_IF";
-	case RING_LOGIC_PER_IP: return "RING_PER_IP";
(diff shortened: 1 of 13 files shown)


src/vma/sock/sockinfo_udp.cpp
```diff
@@ -2637,3 +2584,19 @@ void sockinfo_udp::update_header_field(data_updater *updater)
 		updater->update_field(*m_p_connected_dst_entry);
 	}
 }
+
+void sockinfo_udp::update_dst_entries_ring_logic()
```
liranoz12: dead code ?

src/vma/sock/sockinfo.h
```diff
@@ -288,6 +288,7 @@ class sockinfo : public socket_fd_api, public pkt_rcvr_sink, public pkt_sndr_sou
 
 	virtual int             rx_verify_available_data() = 0;
 	virtual void            update_header_field(data_updater *updater) = 0;
+	virtual void            update_dst_entries_ring_logic() = 0;
```
liranoz12: dead code ?

src/vma/proto/dst_entry.cpp
```diff
@@ -780,3 +780,17 @@ int dst_entry::get_priority_by_tc_class(uint32_t pcp)
 	}
 	return VMA_DEFAULT_ENGRESS_MAP_PRIO;
 }
+
+bool dst_entry::update_ring_alloc_logic(int fd, lock_base& socket_lock, resource_allocation_key &ring_alloc_logic)
```
liranoz12: I see no reason for this function.
We should modify do_ring_migration to public and do the below logic inside ring_alloc_logic_updater::update_field.

src/stats/stats_printer.cpp
```diff
@@ -37,6 +37,8 @@
 #include "vma/util/utils.h"
 #include "vma/util/vma_stats.h"
 #include "vma/lwip/tcp.h"
+#include "vma/vma_extra.h"
+#include "vma/util/sys_vars.h"
```
liranoz12: Can be removed if we will agree to remove p_si_stats->ring_alloc_logic_rx

src/vma/proto/dst_entry.h
```diff
@@ -109,6 +110,7 @@ class dst_entry : public cache_observer, public tostr, public neigh_observer
 		return m_header.set_vlan_pcp(get_priority_by_tc_class(pcp)); }
 	inline header*	get_network_header() { return &m_header;}
 	inline ring*	get_ring() { return m_p_ring;}
+	inline ring_allocation_logic_tx* get_ring_alloc_logic() { return &m_ring_alloc_logic;}
```
liranoz12: unused function

src/vma/sock/sockinfo_tcp.cpp
```diff
@@ -3399,7 +3400,14 @@ int sockinfo_tcp::setsockopt(int __level, int __optname,
 	bool supported = true;
 	bool allow_privileged_sock_opt = false;
 
+	if (m_sock_state <= TCP_SOCK_ACCEPT_READY && __optval != NULL && is_inherited_option(__level, __optname))
```
liranoz12: I dont agree with this change.
Lets say we are passing invalid argument. so each establish socket will reset this option during creation.

src/vma/sock/sockinfo_tcp.cpp
```diff
@@ -3399,7 +3400,14 @@ int sockinfo_tcp::setsockopt(int __level, int __optname,
...
+		m_socket_options_list.push_back(new socket_option_t(__level, __optname,__optval, __optlen));
+
 	if ((ret = sockinfo::setsockopt(__level, __optname, __optval, __optlen)) != SOCKOPT_PASS_TO_OS) {
+		if (ret == SOCKOPT_INTERNAL_VMA_SUPPORT &&
+		    m_sock_state <= TCP_SOCK_ACCEPT_READY && __optval != NULL &&
+		    is_inherited_option(__level, __optname))
+			m_socket_options_list.push_back(new socket_option_t(__level, __optname,__optval, __optlen));
```
liranoz12: two push_back() calls ?

src/vma/sock/sockinfo.cpp
```diff
@@ -107,6 +107,11 @@ sockinfo::sockinfo(int fd):
 	m_p_socket_stats->fd = m_fd;
 	m_p_socket_stats->inode = fd2inode(m_fd);
 	m_p_socket_stats->b_blocking = m_b_blocking;
+	m_p_socket_stats->ring_alloc_logic_rx = m_ring_alloc_log_rx.get_ring_alloc_logic();
+	m_p_socket_stats->ring_alloc_logic_tx = m_ring_alloc_log_tx.get_ring_alloc_logic();
+	m_p_socket_stats->ring_user_id_rx = m_ring_alloc_logic.calc_res_key_by_logic();
+	m_p_socket_stats->ring_user_id_tx =
+			ring_allocation_logic_tx(get_fd(), m_ring_alloc_log_tx, this).calc_res_key_by_logic();
```
liranoz12: User id should be set from setsockopt. we should we update it from the constructor ?

src/vma/util/data_updater.cpp
```diff
@@ -79,3 +79,22 @@ bool header_tos_updater::update_field(dst_entry &dst)
...
+	: data_updater()
+	, m_fd(fd)
+	, m_socket_lock(socket_lock)
+	, m_key(ring_alloc_logic)
+	, m_sock_stats(socket_stats)
+{
+
+}
+
+bool ring_alloc_logic_updater::update_field(dst_entry &dst)
```
liranoz12: Do we want to return true for each and every call ?

src/stats/stats_printer.cpp
```diff
@@ -103,6 +121,14 @@ void print_full_stats(socket_stats_t* p_si_stats, mc_grp_info_t* p_mc_grp_info,
...
 	fprintf(filename, "\n");
 
+	//
+	// Ring Allocation Logic information
+	//
+	fprintf(filename, "- RX: ring alloc logic = %s, ring user ID = %lu\n",
+			ring_logic_str(p_si_stats->ring_alloc_logic_rx), p_si_stats->ring_user_id_rx);
+	fprintf(filename, "- TX: ring alloc logic = %s, ring user ID = %lu\n",
+			ring_logic_str(p_si_stats->ring_alloc_logic_tx), p_si_stats->ring_user_id_tx);
+
```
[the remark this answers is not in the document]
Awik84: ring_id is not necessarily coming from the user setting. It's calculated according to logic and current behavior of the application (if it's per thread than the thread number, if it's core than core id). Sometimes it can also be 0 so why not show it?

src/stats/stats_printer.cpp
```diff
@@ -71,6 +73,22 @@ const char* to_str_socket_type_netstat_like(int type)
...
+	case RING_LOGIC_PER_IP: return "RING_PER_IP";
+	case RING_LOGIC_PER_SOCKET: return "RING_PER_SOCKET";
+	case RING_LOGIC_PER_USER_ID: return "RING_PER_USER_ID";
+	case RING_LOGIC_PER_THREAD: return "RING_PER_THREAD";
+	case RING_LOGIC_PER_CORE: return "RING_PER_CORE";
+	case RING_LOGIC_PER_CORE_ATTACH_THREADS: return "RING_PER_CORE_ATTACH_THREADS";
+	default:
+		break;
+	}
+	return "???";
```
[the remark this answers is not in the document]
Awik84: Right... forgot to remove it.

src/vma/util/vma_stats.h
```diff
@@ -188,6 +189,10 @@ typedef struct {
 	uint32_t                     n_tx_ready_byte_count;
 	socket_counters_t            counters;
 	std::bitset<MC_TABLE_SIZE>   mc_grp_map;
+	ring_logic_t                 ring_alloc_logic_rx;
+	ring_logic_t                 ring_alloc_logic_tx;
```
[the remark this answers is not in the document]
Awik84: But with allowing user to change logic via setsockopt, he can change it to any logic he wants (Not only user_id) why not presenting back to user what is the current logic for that socket?

src/vma/sock/sockinfo_tcp.cpp
```diff
@@ -4613,3 +4618,13 @@ void sockinfo_tcp::update_header_field(data_updater *updater)
 
 	unlock_tcp_con();
 }
+
+void sockinfo_tcp::update_dst_entries_ring_logic()
```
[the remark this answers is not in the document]
Awik84: Yes... will remove.

src/vma/proto/dst_entry.h
Awik84: Fixed

src/vma/sock/sockinfo_tcp.cpp
Awik84: I forgot to remove it. I put it under the next condition.

src/vma/sock/sockinfo.cpp
Awik84: There's always a key/user_id to the ring - whether user set it or we got it from calculating the key value according to the core/thread/fd... This is initializing it to the initial value that is generated according to the default ring allocation logic.

src/vma/util/data_updater.cpp
Awik84: So does the other updaters do... in any case, nobody is looking at the return value (See sockinfo_udp/tcp.cpp at sockinfo_udp::update_header_field

src/stats/stats_printer.cpp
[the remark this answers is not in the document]
liranoz12: But UDP socket inculde more than a single ring. also tcp listen socket which bounded to 0.0.0.0.
we need to talk about it

src/vma/util/vma_stats.h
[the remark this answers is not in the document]
liranoz12: i dont want to add redundant lines to vma_stats output. it makes the output to be messy.
Lets talk about it..

liranoz12:
@Awik84,

1. More dead sections :
```
src/vma/sock/sockinfo.h:291:    virtual void            update_dst_entries_ring_logic() = 0;
src/vma/sock/sockinfo_tcp.h:176:        virtual void update_dst_entries_ring_logic();
src/vma/sock/sockinfo_udp.cpp:2588:void sockinfo_udp::update_dst_entries_ring_logic()
src/vma/sock/sockinfo_udp.h:181:        virtual void update_dst_entries_ring_logic();
```

2. You are updating the statistics only at sockinfo constructor and setsockopt.
What about migration ? what about ring-per-ip - the ip is resolved later...

3. i am ok with monitoring ring id + ring logic - but we should not print them from the default values (0, interface).

4. dst_entry::update_ring_alloc_logic should be removed - use should use ring_alloc_logic_updater.

src/stats/stats_printer.cpp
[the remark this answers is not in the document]
liranoz12: Please move this section to be printed before socket statistics

src/vma/proto/dst_entry.cpp
liranoz12: The idea of data_updater is to avoid adding new functions to dst_entry.
Please use data updater instead of this function.

src/vma/sock/sockinfo.cpp
```diff
@@ -350,6 +405,25 @@ int sockinfo::setsockopt(int __level, int __optname, const void *__optval, sockl
...
+						  setsockopt_so_opt_to_str(__optname),
+						  sizeof(vma_ring_alloc_logic_attr), __optlen);
+					break;
+				}
+			}
+			else {
+				ret = SOCKOPT_NO_VMA_SUPPORT;
+				si_logdbg("SOL_SOCKET, %s=\"???\" - NOT HANDLED, optval == NULL", setsockopt_so_opt_to_str(__optname));
+			}
+			break;
```
liranoz12: Do we want to add support for getsockopt  ?

src/vma/sock/sockinfo.h
```diff
@@ -303,7 +303,10 @@ class sockinfo : public socket_fd_api, public pkt_rcvr_sink, public pkt_sndr_sou
 	bool 			detach_receiver(flow_tuple_with_local_if &flow_key);
 	net_device_resources_t* create_nd_resources(const ip_address ip_local);
 	bool                    destroy_nd_resources(const ip_address ip_local);
-	void			do_rings_migration();
+	void			do_rings_migration(resource_allocation_key &old_key);
+	int			set_ring_attr(vma_ring_alloc_logic_attr *attr);
+	int			set_ring_attr_helper(ring_alloc_logic_attr *sock_attr,
+				  		     vma_ring_alloc_logic_attr *attr);
```
liranoz12: please use single line for set_ring_attr_helper()

src/vma/sock/sockinfo_tcp.cpp
liranoz12: The logic is not good.
It will break non-vma parameters such as SO_TIMESTAMPING/IP_TTL/SO_TIMESTAMP/SO_TIMESTAMPNS.

src/vma/util/vma_stats.h
```diff
@@ -200,6 +205,7 @@ typedef struct {
 		n_rx_ready_pkt_count = n_rx_ready_byte_count = n_rx_ready_byte_limit = n_rx_zcopy_pkt_count = n_tx_ready_byte_count = 0;
 		memset(&counters, 0, sizeof(counters));
 		mc_grp_map.reset();
+		ring_user_id_rx = ring_user_id_tx = 0;
```
liranoz12: Need to earse ring_alloc_logic_rx/ring_alloc_logic_tx

src/vma/proto/dst_entry.cpp
```diff
@@ -623,12 +624,11 @@ void dst_entry::do_ring_migration(lock_base& socket_lock)
 	resource_allocation_key *new_key = m_ring_alloc_logic.get_key();
 	uint64_t new_calc_id = m_ring_alloc_logic.calc_res_key_by_logic();
 	// Check again if migration is needed before migration
-	if (new_key->get_user_id_key() == new_calc_id) {
+	if (old_key.get_user_id_key() == new_calc_id &&
```
liranoz12: Please set new key below m_ring_alloc_logic.calc_res_key_by_logic().
It's very confusing...
Also for rx.

src/vma/sock/sockinfo.cpp
```diff
@@ -188,6 +193,56 @@ int sockinfo::fcntl(int __cmd, unsigned long int __arg)
...
+	return SOCKOPT_INTERNAL_VMA_SUPPORT;
+}
+
+int sockinfo::set_ring_attr_helper(ring_alloc_logic_attr *sock_attr,
+				   vma_ring_alloc_logic_attr *user_attr)
+{
+	if (user_attr->comp_mask & VMA_RING_ALLOC_MASK_RING_PROFILE_KEY) {
+		if (sock_attr->get_ring_profile_key()) {
+			si_logdbg("ring_profile_key is already set and "
+				  "cannot be changed");
```
liranoz12: Why do we want to limit this functionality to a single call ?

src/vma/sock/sockinfo_tcp.cpp
liranoz12: and all the rest of the TCP socket options from sockinfo_tcp...

src/vma/proto/dst_entry.cpp
Awik84: Done.

src/vma/util/vma_stats.h
Awik84: - Fixed the sockopt inheritance issue.
-  How will I know that the logic is user_id to print the ID if I remove ring_user_id_rx/tx ? the ID itself is not enough.

src/vma/sock/sockinfo.h
Awik84: Fixed.

src/vma/sock/sockinfo.cpp
Awik84: Added

Awik84:
Regarding update_ring_alloc_logic - it seems that I'm also dealing with dst_entry's private members when doing the logic change (m_ring_alloc_logic) I will have to change that as well to public (together with do_ring_migration). Perhaps it's worth leaving it as is?

src/stats/stats_printer.cpp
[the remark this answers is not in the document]
Awik84: Done

liranoz12:
_Perhaps it's worth leaving it as is?_
 Ok.

Two last comments:
Please add ring id info to : _void sockinfo::statistics_print(vlog_levels_t log_level /* = VLOG_DEBUG */)_
Regarding getsockopt - why did you define new struct ? we already have vma_ring_alloc_logic_attr which used by setsockopt

liranoz12:
and migration stats before socket statistics...

liranoz12:
bot:retest

PR #796, opened 2019-01-09
issue: 1631600 Cleanup socketxtreme and vmapoll configuration parameters

serglelv:


liranoz12:
@serglelv,

Please remove related ifdef here:
https://github.com/Mellanox/libvma/blob/master/src/vma/util/sys_vars.h#L684:L688

Thanks,

PR #800, opened 2019-01-20
issue: 1647737 Modify Packet Pacing capability calculation

liranoz12:
Packet Pacing should be calculated once for each device and not during
QP creation.

### DIFF
--- a/src/vma/dev/ring_simple.cpp
+++ b/src/vma/dev/ring_simple.cpp
@@ -92,25 +92,9 @@ qp_mgr* ring_eth::create_qp_mgr(const ib_ctx_handler* ib_ctx, uint8_t port_num, 
 	return new qp_mgr_eth(this, ib_ctx, port_num, p_rx_comp_event_channel, get_tx_num_wr(), m_partition);
 }
 
-bool ring_eth::is_ratelimit_supported(struct vma_rate_limit_t &rate_limit)
-{
-#ifdef DEFINED_IBV_EXP_QP_RATE_LIMIT
-	return m_p_qp_mgr->is_ratelimit_supported(m_p_ib_ctx->get_ibv_device_attr(), rate_limit);
-#else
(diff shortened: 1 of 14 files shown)


src/vma/ib/base/verbs_extra.cpp
```diff
@@ -270,6 +270,26 @@ int priv_ibv_query_qp_state(struct ibv_qp *qp)
 	return (ibv_qp_state)qp_attr.qp_state;
 }
 
+int priv_ibv_query_burst_supported(struct ibv_qp *qp, uint8_t port_num)
+{
+#ifdef	DEFINED_IBV_EXP_QP_SUPPORT_BURST
+	if (priv_ibv_modify_qp_from_err_to_init_raw(qp, port_num) == 0) {
+		if (priv_ibv_modify_qp_from_init_to_rts(qp, 0) == 0) {
+			struct vma_rate_limit_t rate = {1000, 100, 100};
```
rafiw: what if it doesn't support PP at all?

src/vma/dev/ring_simple.cpp
```diff
@@ -993,7 +993,7 @@ int ring_simple::get_ring_descriptors(vma_mlx_hw_device_data &d)
 {
 	d.dev_data.vendor_id = m_p_ib_ctx->get_ibv_device_attr()->vendor_id;
 	d.dev_data.vendor_part_id = m_p_ib_ctx->get_ibv_device_attr()->vendor_part_id;
-	if (vma_is_packet_pacing_supported(m_p_ib_ctx->get_ibv_device_attr())) {
+	if (m_p_ib_ctx->is_packet_pacing_supported()) {
```
rafiw: please also verify burst is on

liranoz12: I will do it as a separate PR
:+1:

PR #806, opened 2019-01-31
issue: 1599446 Fix RX TAP migration failure for 3t connections

liranoz12:
Filter node id selection for 3t-rules was based on the port value.
Hence, during migration, two TAP devices had the exact same rule.
This commit modifies this logic to match the 5t-rules logic.


liranoz12:
bot:retest

PR #810, opened 2019-02-06
issue: 1677457 Add methods mask to VMA extra api

liranoz12:
This adds additional mask to the extra API header so the user would
be able to query which methods were initialized by VMA as part of
vma_get_api().


liranoz12:
bot:retest

PR #871, opened 2019-12-17
issue: 2016629 Add setsockopt to shutdown socket's RX side

alexvm:
Add new VMA socket option that closes the RX side of the socket.
This is needed to avoid steering rules creation for TX ONLY sockets.
Adding steering entries in this case hurts performance, especially for big scale cases.


alexvm:
setsockopt(SO_VMA_SHUTDOWN_RX) should be called AFTER bind().

PR #1116, opened 2025-03-19
issue: 4337575 Init VMA in vma_thread_offload()

pasis:
## Description

vma_get_api()->thread_offload() can be called before the 1st socket() syscall. In this case VMA is in uninitialized state and the thread_offload() call fails with -1.

Call DO_GLOBAL_CTORS() in the vma_thread_offload() to guarantee the initialized state in all scenarios.

Until VMA 9.8.70, the initialization was done in the vma_get_api() call, however, it was removed due to external allocator support which requires uninitialized state to guarantee buffers allocation with the configured allocator.

##### What
Fix vma_thread_offload() failure is called before the 1st socket() syscall. 

##### Why ?
Bugfix

## Change type
- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/vma/sock/sock-redirect.cpp
+++ b/src/vma/sock/sock-redirect.cpp
@@ -609,6 +609,8 @@ int vma_add_conf_rule(const char *config_line)
 extern "C"
 int vma_thread_offload(int offload, pthread_t tid)
 {
+	DO_GLOBAL_CTORS();
+
 	if (g_p_fd_collection) {
 		g_p_fd_collection->offloading_rule_change_thread(offload, tid);
 	} else {


igor-ivanov (approved):

src/vma/sock/sock-redirect.cpp
```diff
@@ -609,6 +609,8 @@ int vma_add_conf_rule(const char *config_line)
 extern "C"
 int vma_thread_offload(int offload, pthread_t tid)
 {
+	DO_GLOBAL_CTORS();
```
galnoam, line 612: What if we call vma_thread_offload multiple times?

PR #1170, opened 2026-01-07
issue: 4504533 Add warning when BlueFlame setup fails

BasharRadya:
Add a warning message when BlueFlame (BF) is requested via handle_bf but cannot be set up due to hardware or environment limitations.

## Description
Please provide a summary of the change.

##### What
_Subject: what this PR is doing in one line._ 

##### Why ?
_Justification for the PR. If there is existing issue/bug please reference._

##### How ?
_It is optional but for complex PRs please provide information about the design,
architecture, approach, etc._

## Change type
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

## Check list

### DIFF
--- a/src/vma/dev/qp_mgr_eth_mlx5.cpp
+++ b/src/vma/dev/qp_mgr_eth_mlx5.cpp
@@ -56,6 +56,7 @@ static inline uint64_t align_to_WQEBB_up(uint64_t val)
 static bool is_bf(struct ibv_context *ib_ctx)
 {
 	char *env;
+	bool bf_supported = false;
 
 	/* This limitation is done for RM: 1557652, 1894523, 1914464, 2069198 */
 	if (safe_mce_sys().hypervisor != mce_sys_var::HYPER_NONE) {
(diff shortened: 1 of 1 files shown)


src/vma/dev/qp_mgr_eth_mlx5.cpp
```diff
@@ -84,11 +85,16 @@ static bool is_bf(struct ibv_context *ib_ctx)
...
-			return true;
+			bf_supported = true;
 		}
 #endif /* MLX5DV_UAR_ALLOC_TYPE_BF */
 	}
-	return false;
+
+	if (safe_mce_sys().handle_bf && !bf_supported) {
+		qp_logwarn("Unable to Setup BlueFlame, falling back to doorbell method");
+	}
```
pasis, line 95: I think we should be more verbose here and explain the impact. Instead of technical details about doorbell method, we should describe that this is not critical, however, this can affect performance.

BasharRadya: "		__log_warn("Unable to initialize BlueFlame; falling back to doorbell method. "
			   "Small packet latency performance may be reduced.");"
			   
			   better?

pasis: we don't have to mention doorbell method at all. This is obvious for us and extra technical information for users. Also all messages are affected, because BF removes 1 PCI transaction for larger messages.

pasis (approved):

